Issue #11, opened 2020-05-20
Sync with Upstream


knapply:
Per the [contributing guidance](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md#submitting-a-pull-request), here is a planned PR. 

The main motivation (beyond keep pace) is access to `simdjson::dom::object` and `::array` sizes for better preallocation.

* [Proposed Branch](https://github.com/knapply/rcppsimdjson/tree/cran/)
* Changes:
  + [inst/include/simdjson.h](https://github.com/knapply/rcppsimdjson/blob/cran/inst/include/simdjson.h) 
    + Synced with upstream
  + [inst/include/simdjson.cpp](https://github.com/knapply/rcppsimdjson/blob/cran/inst/include/simdjson.cpp) 
    + Synced with upstream
  + [src/simdjson_example.cpp](https://github.com/knapply/rcppsimdjson/tree/cran/src/simdjson_example.cpp) (parseExample) 
    + Added [simdjson namespace to `.get<dom::array>()`](https://github.com/knapply/rcppsimdjson/blob/0015dda2d80398b05fe47bd3014c72ad0369d425/src/simdjson_example.cpp#L40) so that it still compiles
  + Updated [ChangeLog](https://github.com/knapply/rcppsimdjson/blob/cran/ChangeLog)

I'm happy to update [NEWS.Rd](https://github.com/knapply/rcppsimdjson/blob/cran/inst/NEWS.Rd)  if the following conforms...

```
...

\section{Changes in version 0.0.5 (2020-xx-xx)}{
  \itemize{
    \item Synced with upstream (Brendan in \ghpr{x}
    \item Updated example \code{parseExample} to API changes.
  }
}

...
```

eddelbuettel:
That's prize winning material :)   Appreciate the attention to detail.   Following the initial few releases which ironed out some things (though eg the `stderr` reverted as your noticed) updates should be "cheap".  Because we don't have much in terms of usage there isn't much in terms of test coverage either.

knapply:
Beyond being insanely fast, the simdjson API is _really_ slick. More functionality, tests, and users seem inevitable.

I wanted to sync before tackling anything else (including giving @dcooley a second set of eyes for `from_json()`), but I definitely have some ideas that I've been playing with elsewhere. I'll bring those up separately when time permits.

eddelbuettel:
[...]

PR #12, opened 2020-05-20
Sync with Upstream

knapply:
Referencing #11


eddelbuettel (approved):
Thanks for the update!

PR #32, opened 2020-07-05
Adding fparse, fload and more polish

eddelbuettel:


eddelbuettel (commented):
Rockin!  Apparently I can't approve this as I am listed as the PR author :roll_eyes:

knapply:
Windows success: https://builder.r-hub.io/status/RcppSimdJson_0.0.6.4.tar.gz-f6959dcc7db5abb5e35717c3ed3299d1

Give me a few. I may have a fix for the CRAN Mac.

It [throws a fit over `std::optional`](https://builder.r-hub.io/status/RcppSimdJson_0.0.6.4.tar.gz-97f536bd324db95ca4510203d8043497#L367) so it's exaggerating it's c++-17-ness pretty hard, but I think I can grab it from `<experimental/optional>`.

_If_ I can't fix it, how do you feel about initaiting option `#ifdef DEAD_MEAT` on that system?

eddelbuettel:
I really like `#if defined(DEADMEAT)` as a general fsck you for systems where we have no hope whatsover (hello, Solaris, no need to blush here...) but would prefer to avoid it if we can.  Just because so many other devs use crapOS it would be sad to exclude them, especially now that we have windoze covered.

knapply:
Winner for Mac: https://builder.r-hub.io/status/RcppSimdJson_0.0.6.4.tar.gz-db5eb4365e08415bc961fc1dc52ac2a4

Just checking I'm not breaking anything elsewhere and making the code formatting less... "weak" 🤨

knapply (approved):

Issue #33, opened 2020-07-08
unlink woes


knapply:
https://twitter.com/michael_chirico/status/1280675491448410113

"just poking around so I may be wrong, but it looks like that unlink should be diagnosis$input[[i]]?"

![image](https://user-images.githubusercontent.com/26281249/86864650-691cd480-c082-11ea-89f5-d50bac4132ed.png)

Once upon a time it was only URLs or only files. and `unlink()` would only affect downloaded files. That's no longer the case.

eddelbuettel:
That was too short-hand for me.  Maybe copy the tweet text time.

Bugs happen. I created millions in my day.  Life goes on.  Reasons for 0.1.1 :)

eddelbuettel:
Thanks for the edit. I got there eventually. (It doesn't help that Chrome+Twitter, the combo of adblockers I have, don't directly send me to the tweet (high _wtf_ factor here) so I always have to go a send time in incognito mode.  Ahh the 1st world problems we have.  Now you have know why I whine about directly quoting ;-)

BTW looks like you pushed on "empty" ie "unchanged" branch in issue/33.  On purpose?

knapply:
Yes, it was intentional.

eddelbuettel:
Gotta love a good baseline to compare to :)

eddelbuettel:
Done in #34

PR #34, opened 2020-07-08
fix non-temp file deletion bug

knapply:
Addressing https://github.com/eddelbuettel/rcppsimdjson/issues/33

... the formatting change didn't quite stick from the looks of it, but that should be ironed out.

### DIFF
--- a/R/fload.R
+++ b/R/fload.R
@@ -1,13 +1,13 @@
 .file_extension <- function(x) {
   base_name <- basename(x)
-  
+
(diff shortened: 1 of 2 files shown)


knapply:
This is also as good a time as any to confirm CRAN plays along with downloading a file.

eddelbuettel (approved):
Looks good.  Had to turn 'ignore whitespace' on.  Not a biggie though.

knapply:
I must've checked off that box _after_ formatting the file or something.

Fortunately, the bug only bites if you're reading from both local/remote files simultaneously, but gnarly nonetheless!

PR #61, opened 2020-12-09
add GitHub Actions

eddelbuettel:
Travis CI keeps talking about unlimited use for open source projects yet my runs are currently stalled -- not nice.

So this simple PR turns on GitHub Action with a build matrix just like the one we had before. 

No code, documentation or other changes.


eddelbuettel:
The code coverage change seems spurious.  From the look at the `.travis.yml` we were running it for each of the four gcc/g++ setups (oops, wasteful) whereas I now run it only for gcc-10/g++-10.  That should not affect things but we can sort that out another time.

PR #65, opened 2021-02-10
(issue #64) add `always_list` parameter (fparse/fload)

### LINKED ISSUE #64: Add 'enforce_lapply' or similar to fparse/fload
It is a great feature of the package that one does not need to call `lapply()` on character vectors since `fparse\fload` are already vectorized. However, this also means that the output type depends on whether the input contains a single element or multiple elements, e.g.:

```
inp2 <- as.character(1:2)
RcppSimdJson::fparse(inp2)
# [[1]]
# [1] 1
# 
# [[2]]
# [1] 2

inp1 <- as.character(1)
RcppSimdJson::fparse(inp1)
# [1] 1
```

So to be on the safe side, one has to check if the input vector is of length 1, and if yes, one has to wrap the result in a list. E.g.:

```
safe_fparse <- function(json, ...) {
  out <- RcppSimdJson::fparse(json, ...)
  if (length(json) == 1L) out <- list(out)
  out
}
safe_fparse("1")
# [[1]]
# [1] 1
```

What about adding an argument to fparse/fload which **enforces** that each element of the input json is wrapped in a list?

knapply:
Resolves #64.

### DIFF
--- a/R/fload.R
+++ b/R/fload.R
@@ -63,6 +63,7 @@ fload <- function(json,
                   max_simplify_lvl = c("data_frame", "matrix", "vector", "list"),
                   type_policy = c("anything_goes", "numbers", "strict"),
                   int64_policy = c("double", "string", "integer64", "always"),
+                  always_list = FALSE,
                   verbose = FALSE,
                   temp_dir = tempdir(),
                   keep_temp_files = FALSE,
@@ -85,6 +86,9 @@ fload <- function(json,
     }
     if (!.is_scalar_lgl(query_error_ok)) {
         stop("`query_error_ok=` must be either `TRUE` or `FALSE`.")
+    }
+    if (!.is_scalar_lgl(always_list)) {
+        stop("`always_list=` must be either `TRUE` or `FALSE`.")
     }
     if (!.is_scalar_lgl((verbose))) {
         stop("`verbose=` must be either `TRUE` or `FALSE`.")
@@ -176,7 +180,7 @@ fload <- function(json,
     }
 
     # load =====================================================================
-    .load_json(
+    out <- .load_json(
         json = input,
         query = query,
         empty_array = empty_array,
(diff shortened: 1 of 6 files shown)


eddelbuettel (approved):
Looks good as usual!  I can deal with the codecov nag separately.

I think I would have put the new named argument at the end by some time honoured convention though ...

knapply:
> Looks good as usual! I can deal with the codecov nag separately.
> 
> I think I would have put the new named argument at the end by some time honoured convention though ...

The codecov should be fixed momentarily.  The rationale behind the argument order is that the arguments that are applicable to both `fparse()`/`fload()` go before the ones that are only applicable to `fload()`, but I can change that if desired.

eddelbuettel:
Makes sense, I didn't pick up on that.

R/fload.R
```diff
@@ -196,7 +196,7 @@ fload <- function(json,
     )
 
     if (always_list && length(json) == 1L) {
-        `attr<-`(list(out), "names", names(json))
+        `names<-`(list(out), names(json))
```
tdeenes, line 199: `setNames()` is the more idiomatic way to set names for an object.

tdeenes:
Thank you very much for the open-mindedness and the quick implementation!

R/fload.R
knapply: I suppose, but it's the exact same behavior without needing to add `{stats}` to the `DESCRIPTION`'s `Imports:`.

eddelbuettel: Maybe, maybe not. `setNames()` is from `stats`, reading help suggests it is "just a short form" for assigning and naming.  I am cool with `base:names()`.  Can't argue with `base` even if there are shortcuts :)

tdeenes: Indeed, I forgot about the fact that `setNames` comes from **stats**. Sorry for the noise :)

PR #67, opened 2021-02-14
sanitize path from ? upsetting file ops on windows (closes #66)

### LINKED ISSUE #66: pathname with `?` in windows  blocking download
Ref:  https://github.com/eddelbuettel/td/issues/1

The first url fails because the constructed path name intended for `destfile` in `download.file()` 
includes `?` and therefore it is not valid in windows. The issue occurs on the first url  because the last path separator is before  `?` , so when the `basename()` is called inside `.prep_input()` the `?` remains. See  below.

``` r
qry1 <- "https://api.twelvedata.com/time_series?symbol=VTI&interval=1min&apikey=demo&source=docs"
qry2 <- "https://api.twelvedata.com/time_series?symbol=EUR/USD&interval=1min&apikey=demo&source=docs"

RcppSimdJson:::.prep_input(qry1, temp_dir = tempdir(),
                           compressed_download = FALSE,
                           verbose = FALSE)
#> Error in download.file(url = .url, destfile = .destfile, method = .method, : cannot open destfile 'C:\Users\Eluvias\AppData\Local\Temp\RtmpCoLe99\time_series?symbol=VTI&interval=1min&apikey=demo&source=docs39644039333f', reason 'Invalid argument'

RcppSimdJson:::.prep_input(qry2, temp_dir = tempdir(),
                           compressed_download = FALSE,
                           verbose = FALSE)
#>                                                                                                             input
#> 1 C:\\Users\\Eluvias\\AppData\\Local\\Temp\\RtmpCoLe99\\USD&interval=1min&apikey=demo&source=docs396472246445
#>   url_prefix file_ext is_from_url is_local_file_url is_remote_file_url
#> 1   https://                 TRUE             FALSE               TRUE
```
### Part of destfile
```r
# basename is called inside .prep_input()
basename(qry1)
#> [1] "time_series?symbol=VTI&interval=1min&apikey=demo&source=docs"

basename(qry2)
#> [1] "USD&interval=1min&apikey=demo&source=docs"
```

eddelbuettel:
We have already seen (in #66) that this is a minimally complete fix.  

But while we're at it, should we substitute more?  Can you think of another symbol or two?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2021-02-14  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Version, Date): Roll version and date
+
+	* R/utils.R (.prep_input): Sanitize filename for Windows
+
 2021-02-12  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Release 0.1.4
@@ -7,7 +13,7 @@
 
 2021-02-10  Dirk Eddelbuettel  <edd@debian.org>
 
-        * DESCRIPTION (Version, Date): Roll version and date
+         * DESCRIPTION (Version, Date): Roll version and date
 
 2021-02-10  Brendan Knapp  <brendan.g.knapp@gmail.com>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,10 @@
 
 	* R/utils.R: Add #nocov tags
 	* R/validateJSON.R (validateJSON): Idem
+	* inst/include/RcppSimdJson/utils.hpp (decompress): Idem
+	* inst/include/RcppSimdJson/deserialize/dataframe.hpp: Idem
+	* inst/include/RcppSimdJson/deserialize/vector.hpp: Idem
+	* inst/include/RcppSimdJson/deserialize.hpp: Idem
 
 2021-02-14  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2021-02-20  Dirk Eddelbuettel  <edd@debian.org>
+
+	* R/utils.R: Add #nocov tags
+	* R/validateJSON.R (validateJSON): Idem
+
 2021-02-14  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll version and date
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: RcppSimdJson
 Type: Package
 Title: 'Rcpp' Bindings for the 'simdjson' Header-Only Library for 'JSON' Parsing
-Version: 0.1.4
-Date: 2021-02-12
+Version: 0.1.4.1
+Date: 2021-02-14
 Author: Dirk Eddelbuettel, Brendan Knapp
 Maintainer: Dirk Eddelbuettel <edd@debian.org>
 Description: The 'JSON' format is ubiquitous for data interchange, and the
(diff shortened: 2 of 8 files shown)


knapply:
This works, but there's a bit to unpack here.

It seems (some of) the options are:
- handle these oddities one-by-one as they appear
- ditch the current approach entirely and go straight to `{curl}` (don't create files at all, just load the data into memory)
- reimplement `fs::path_sanitize()` (being compatible with it is probably wise).  I have no idea what license issues come into play here (there's only so many ways to do that), but it could look something like this:


```r
.path_sanitize <- function(file_path, replacement = "") {
    stopifnot(is.character(file_path) && !anyNA(file_path))
    case_sensitivity <- c(
        "[/\\?<>\\:*|\":]" = FALSE,
        "[[:cntrl:]]" = FALSE,
        "^[.]+$" = FALSE,
        "^(con|prn|aux|nul|com[0-9]|lpt[0-9])([.].*)?$" = TRUE,
        "[. ]+$" = FALSE
    )
    target_patterns <- names(case_sensitivity)
    for (i in seq_along(target_patterns)) {
        file_path <- gsub(pattern = target_patterns[[i]], 
                          replacement = replacement, 
                          x = file_path, 
                          ignore.case = case_sensitivity[[i]])
    }
    file_path <- substr(file_path, 1L, 255L)
    if (replacement == "") {
        file_path
    } else {
        path_sanitize(file_path, "")
    }
}
```


Unfortunately, C++17's `<filesystem>` seems to require some linking hackery for GCC<8, so just relying on that probably isn't realistic. 


But there's a disconnect between this use case and `fload()`'s original goals (which obviously drove its design).  `fload()` being able to load _files_ from a URL was really only meant to make downloading files convenient, but the real issue here is that this isn't downloading a file (there's no real file name).  Obviously use patterns expanded since we started (I even did an API query in the examples), but I feel like we should avoid reinventing the wheel internally when we really just wanted a wicked fast JSON parser.

eddelbuettel:
Yes/no/maybe.  One could also argue that the real issue may be constructing a file name by trusting a string we're given, and we now found that in some cases this can in fact bite us. So would a fuller sanitization be to just call `tempfile()` and simply not feed it a pattern which may contain indestigable components?

knapply:
That should work, the only issue I can think of is regarding this current behavior:

``` r
library(RcppSimdJson)

a_url <- "https://api.github.com/users/lemire"
multiple_urls <- c(
    a_url,
    "https://api.github.com/users/eddelbuettel",
    "https://api.github.com/users/knapply",
    "https://api.github.com/users/dcooley"
)
fload(multiple_urls, query = "/name")
#> $lemire
#> [1] "Daniel Lemire"
#> 
#> $eddelbuettel
#> [1] "Dirk Eddelbuettel"
#> 
#> $knapply
#> [1] "Brendan Knapp"
#> 
#> $dcooley
#> [1] "Dave"
```

Right now, the `names()` use the "file name".  I suppose a way around that would be to check file names for illegal characters and just replace any strings that contain them with `tempfile()` (which is at least simpler to maintain internally).  We could also just ditch this behavior.

eddelbuettel:
Could we in that case not set names?   (I understand that I am pushing your spec a little with my (ab)use from package `td`. I just checked and `td` is equally happy to use `format=csv` but as, always, "that's why we can't have nice things" this then prohibits batch use over multiple symbols. Gaa.)

The name settings also had a minor hoopla here in that we set the key directly in the URL as a 'transient' sacred payload and then we ... promptly get it printed back as the tempfile name. Whoops.  It doesn't really spill but there are other cases where the 'URL  defines filename' use has minor warts.

(And we may also decide that there is no clear winner either way and just keep the one-off filter I added.  We can always expand as needed.  Keeping it simple(r) ? )

eddelbuettel:
Will merge this now and plan to ship to CRAN 'soon' (say, tomorrow?) to plug the hole we have on Windows.  That is of moderate urgency,  A better more thorough sanitizer is a good idea but we don't have to rush it.

Issue #78, opened 2023-01-19
Export `.deserialize_json()`


shikokuchuo:
Please consider exporting `.deserialize_json()`, as `fparse()` performs a lot of safety checks, detracting needlessly from performance. Preferably with the 'simplify_to' argument defaulting to 3L, again for best performance. Thanks.

eddelbuettel:
I am all ears.  Can you possibly sketch it out in a PR (that we maybe iterate over) ?

eddelbuettel:
You can already call it.  Isn't that good enough?

```r
> j <- jsonlite::toJSON(data.frame(a=c(42L, 88L), b=c("brown", "fox")))
> j
[{"a":42,"b":"brown"},{"a":88,"b":"fox"}] 
> RcppSimdJson:::.deserialize_json(j)
   a     b
1 42 brown
2 88   fox
> 
```

shikokuchuo:
I have created the PR for your perusal.

The overhead from using `:::` ...
+ it would allow me to call it from a CRAN package (although that is not a priority)

eddelbuettel:
(See discussion at your PR #79)

dcooley:
> it would allow me to call it from a CRAN package (although that is not a priority)

If this is your goal, you can `LinkTo` this library, include the [`RcppSimdJson/deserialize.hpp`](https://github.com/eddelbuettel/rcppsimdjson/blob/master/inst/include/RcppSimdJson/deserialize.hpp) and write your own version of `deserialize` in your own package?

shikokuchuo:
@dcooley that's interesting I had not considered something like that. I have my own fork for my own use. I just thought other CRAN users may want to benefit from the same tweaks.

eddelbuettel:
1. It's `LinkingTo` and it exposes header files.  Actually exporting (at the _foreign function interface_ C level calls across packages is more powerful.  We could even export the internal C++ function with its `SEXP` only C interface).
2. But we still have not seen _any_ rationalization for the rumoured speed increase and what your usage pattern: few large JSON blobs?  Many small ones?
3. You still can _today_ workaround the `:::` limitation in CRAN package checks. Many packages do.

eddelbuettel:
I do not think we are spending time on worthwhile initiatives here:

```r
> jsoninputfile <- system.file("jsonexamples", "twitter.json", package="RcppSimdJson")
> jsoninput <- paste(readLines(jsoninputfile), collapse="")
[...]

PR #79, opened 2023-01-19
export .deserialize_json()

shikokuchuo:
PR for issue #78 to export `.deserialize_json()` with the most performant defaults.


.gitignore
```diff
@@ -2,8 +2,10 @@
 .Rhistory
 .RData
 .Ruserdata
+*.Rproj
```
eddelbuettel, line 5: Please delete.  We don't use projects.

.gitignore
```diff
@@ -2,8 +2,10 @@
...
 src/*.o
 src/*.so
 src/*.dll
+src/Makevars
```
eddelbuettel, line 9: Please delete.  `cleanup` removes that already.

NAMESPACE
```diff
@@ -1,4 +1,5 @@
 useDynLib(RcppSimdJson, .registration=TRUE)
 exportPattern("^[[:alpha:]]+")
+export(.deserialize_json)
```
eddelbuettel, line 3: No.  Functions starting with a dot are internal and by default not exported.  We would not export a function with a dot.

One _could_ consider remaning the helper function to make it visible.  @knapply and I once decided to have it hidden (and have the extra safeties in a user-facing function).

R/RcppExports.R
```diff
@@ -1,7 +1,7 @@
 # Generated by using Rcpp::compileAttributes() -> do not edit by hand
 # Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
 
-.deserialize_json <- function(json, query = NULL, empty_array = NULL, empty_object = NULL, single_null = NULL, parse_error_ok = FALSE, on_parse_error = NULL, query_error_ok = FALSE, on_query_error = NULL, simplify_to = 0L, type_policy = 0L, int64_r_type = 0L) {
+.deserialize_json <- function(json, query = NULL, empty_array = NULL, empty_object = NULL, single_null = NULL, parse_error_ok = FALSE, on_parse_error = NULL, query_error_ok = FALSE, on_query_error = NULL, simplify_to = 3L, type_policy = 0L, int64_r_type = 0L) {
```
eddelbuettel, line 4: I am not sure I am in favour of sudden and hidden behavior changes.

Maybe (and that is a big maybe) we can add a new user-facing function which defaults to calling with `simplify_to=3L`.

src/deserialize.cpp
```diff
@@ -12,7 +12,7 @@ SEXP deserialize(SEXP       json,
                  SEXP       on_parse_error = R_NilValue,
                  const bool query_error_ok = false,
                  SEXP       on_query_error = R_NilValue,
-                 const int  simplify_to    = 0,
+                 const int  simplify_to    = 3,
```
eddelbuettel, line 15: As above.

NAMESPACE
shikokuchuo: The point is that it would not be for 'general consumption' but for 'expert' use. There are often so-called technical utility functions starting with a '.'

shikokuchuo: I am suggesting this as the point of RcppSimdJson is that it is fast. It seems a waste to not unleash it's full potential!

eddelbuettel:
Can you motivate the new function interface with timing evidence?  As I mentioned and showed in [the comment to your issue](https://github.com/eddelbuettel/rcppsimdjson/issues/78#issuecomment-1397653535) you can already call it (though not from another package, at least not directly, but you could [play tricks](https://github.com/rstudio/reticulate/blob/main/R/utils.R#L49-L51)).

The main question is: what speed difference are we talking about here?

R/RcppExports.R
shikokuchuo: As it is not currently exported, there is no behavioural change. `fparse()` always calls it with the arguments defined.

eddelbuettel:
Crikey NO:

image

The change you wanted to make in your PR was to undo the addition of two lines we do not need, NOT to delete the file we do need.  `git restore` may do what you need.

shikokuchuo:
Sorry I think we are back to where we were. Failed with the github GUI as I have multiple forks...

eddelbuettel:
`git` is forgiving.  You can `git reset --hard` to a know good place and start over.  There is not that much worth keeping in the PR and as argued in https://github.com/eddelbuettel/rcppsimdjson/issues/78#issuecomment-1397777088 I am not sure we are spending our time wisely here.

